PR #6, opened 2017-03-11
Stop trapping exceptions -- until this is explained better

pb-cdunn:
Unfortunately, `dispatch(main)` is trapping exceptions from main. That is a big problem. Why not let exceptions propagate? Is there something in particular that needs to be trapped?

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -312,7 +312,7 @@ macro dispatchGen*(pro: typed, cmdName: string="", doc: string="",
         `callPrs`
         `callWrapd`
       except:
-        discard
+        raise
   when defined(printDispatch): echo repr(result)  # maybe print generated code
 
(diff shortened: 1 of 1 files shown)


c-blake:
This was a quick hack way to have the "help" option case both not exit-non-zero and not call the wrapped function.  Probably the right fix is to raise and catch a very specific exception and let any other sorts of exceptions propagate.

c-blake:
Take a look at the behavior of my main branch.  It still uses a raise/except as a control flow trick, but I just define a local exception type.  So, any other exceptions should propagate up.  If you like how it works, I'll consider this PR thread/issue resolved.  Thanks.

c-blake:
I am pretty sure this addressed your complaint.  Re-open otherwise.

pb-cdunn:
Yes, 571207d846711ee641e41b785352ef9ed529446f solves the problem. Thanks.

PR #32, opened 2018-08-28
Refactor: replace nil strings with "", nil seqs with @[]

kaushalmodi:


c-blake:
Well, testing against the empty value and the length is redundant.  The reason the checks against ``nil`` were there was from a time before ``len`` on ``nil`` would return 0 anyway.  Give me a minute and I'll push my whole set of changes and you can see what I did.  Mostly everything is fine except there are some ``argParse`` situations where I was really distinguishing between the user providing an empty string vs. not providing anything at all.  I'm not quite sure how those should be handled in this new no-``nil`` world order.

kaushalmodi:
>  I'm not quite sure how those should be handled in this new no-nil world order.

Use options.

kaushalmodi:
Examples:

- https://ptpb.pw/CwPr/nim
- https://scripter.co/notes/nim/#options (see how the classic *nil*, now `None` string is distinguished from an empty string)

c-blake:
Yeah, but it's ``parseopt3`` that knows/decides whether there is an empty value vs. no value at all.  So, this would change that API, making it even more different from ``parseopt2`` in the stdlib.  I've been trying to keep things mostly backward compatible for people.

kaushalmodi:
>  I've been trying to keep things mostly backward compatible for people.

Though it's easy to decide which package versions using nimble. For people using nim 0.18.0, they can set the dependence to use cligen <= 0.9.15, and for newer nim version, use version > 0.9.15.

c-blake:
I think you misunderstood my meaning.  I meant keep ``parseopt3`` backward compatible with ``parseopt2`` (but with more features) to ease any code migration/etc.  If we make ``OptParser.val`` an option type in ``parseopt3``, but it stays a regular string in ``parseopt2`` then that is more fundamentally different than current differences..maybe not that important, but distinguishing a missing val from an empty val is also a rare/error case and maybe not so important either.  So, it's a contest of priorities..maybe with no bad answer (or any great answer..).

Issue #88, opened 2019-02-25
argcvt.nim(5, 31) Error: cannot open file: cligen/textUt


pb-cdunn:
```
foo.nim(9, 11) template/generic instantiation from here
../repos/cligen/cligen.nim(673, 14) template/generic instantiation from here
../repos/cligen/cligen/argcvt.nim(5, 31) Error: cannot open file: cligen/textUt
```
`cligen` is a symlink to `../repos/cligen`, and `../repos/cligen` is where I check-out your sub-modules.

foo.nim:
```
import "cligen/cligen"
```
`cligen.nim`:
```
  1 import macros, tables, cligen/parseopt3, strutils, os
```
`argcvt.nim`:
```
  5 import strformat, sets, cligen/textUt, parseopt3
```
The fix is to use relative imports. (Maybe this is some other fix?) I'll post a PR ...

c-blake:
There was a time when `parseopt3`, `argcvt`, and `cligen` all lived up a level but `textUt` did not which is where that "cligen/" came from.  Back in 2015 when I started, I had been hoping that nimble might grow to impose less project structure, like letting you declare multiple exported modules and also perhaps hoping the stdlib might take `parseopt3` leaving just `cligen` and `argcvt` on equal footing (with the idea that `argcvt` might be usable in other contexts, but `cligen` wouldn't make sense without `argcvt` and without proliferating packages).  When neither of those seemed likely to ever happen and people complained about nimble warnings Re project structure, I just put everything under `cligen/` except `cligen.nim`.  I missed the `cligen/textUt` change I should have made back then.  Thanks for catching this.

Let me know if this doesn't fix your problem.

PR #89, opened 2019-02-25
Fix 88

pb-cdunn:
Background here: https://github.com/jangko/msgpack4nim/issues/30

Also note araq's comment here: https://forum.nim-lang.org/t/4681#29195 
But I don't think that problem is related.

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -1,4 +1,4 @@
-import macros, tables, cligen/parseopt3, strutils, os
+import macros, tables, ./cligen/parseopt3, strutils, os
 type HelpOnly*    = object of Exception
 type VersionOnly* = object of Exception
 type ParseError*  = object of Exception
@@ -207,7 +207,7 @@ proc next*(x: openArray[ClParse], stati: set[ClStatus], start=0): int =
   for i, e in x:
     if e.status in stati: return i
 
-include cligen/syntaxHelp
+include ./cligen/syntaxHelp
 
 macro dispatchGen*(pro: typed{nkSym}, cmdName: string = "", doc: string = "",
  help: typed = {}, short: typed = {}, usage: string=dflUsage,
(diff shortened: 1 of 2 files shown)


pb-cdunn:
I see your fix here: https://github.com/c-blake/cligen/commit/9c2d41955b63be7a9802aad4381a374fc6859638

Trying that  instead now ...

pb-cdunn:
(This was a fix for #88.)

c-blake:
For what it's worth, at least according to the docs for module search paths, that `"./"` is not syntactically necessary, though it may represent useful meaning to a human reader of the code emphasizing locality of the module.  I don't set up symlinks like you, though, and it's possible there are bugs.  So, let me know if it breaks without the `"./"`.

pb-cdunn:
Could you link the docs on search-path? I was looking for that but didn't find it.

c-blake:
Well, it's in `doc/nimc.rst` in the distribution.  In-file-search for "Search path handling" to get to the section.

c-blake:
I agree that the "Collective imports from a directory" section in the manual that gets HTML-ified could use a few sentences about this (or maybe a new section just before it?). I find it hard to imagine that Araq would reject such a PR if you sent one, though you may have to iterate on the exact language.  Also, I think it's possible to `import` out of band from the `.nim` codebase from within `nim.cfg` (and probably `config.nims`).  `tests/import_in_config` and `tests/modules` show that, but I doubt it's documented anywhere.  Also, I believe the search path for `include` works just like `import` but that may not be mentioned anywhere, exactly.  Also, I wouldn't be shocked if there was some `pragma` that could update the search path in Nim code, but I don't know of one specifically.  You could also try to just email miran as he's been doing a lot of doc updating.

c-blake:
Oh, and https://nim-lang.org/docs/nimc.html#compiler-usage-search-path-handling is a bona fide link you can just click if that helps.

pb-cdunn:
Oh! In the *compiler* manual. Of course! Thanks.

c-blake:
You're welcome.  There really should be at least a _pointer_ to that section of the compiler manual from the language manual somewhere around "Collective imports".  That is also talking about subdir structure of to-be-imported stuff.

PR #104, opened 2019-05-16
Fix detection of stable

genotrance:
cligen fails on 0.19.6

```
~/.nimble/pkgs/cligen-0.9.26/cligen/textUt.nim(7, 13) Error: cannot open file: std/wordwrap
```


c-blake:
(Also, I will probably punch a release of `cligen` soon so that non-HEAD `cligen` dependency can work.)

PR #121, opened 2019-10-01
Prepare for csize to be changed to uint

Clyybber:
See https://github.com/nim-lang/Nim/pull/12321 and https://github.com/nim-lang/Nim/issues/12187

### DIFF
--- a/cligen/sysUt.nim
+++ b/cligen/sysUt.nim
@@ -6,7 +6,7 @@ proc `:=`*[T](x: var T, y: T): T =
 proc findUO*(s: string, c: char): int {.noSideEffect.} =
   proc memchr(s: pointer, c: char, n: csize): pointer {.importc:"memchr",
                                                         header:"<string.h>".}
-  let p = memchr(s.cstring.pointer, c, s.len)
+  let p = memchr(s.cstring.pointer, c, s.len.csize)
   if p == nil: -1 else: (cast[uint](p) - cast[uint](s.cstring.pointer)).int
 
 proc delete*(x: var string, i: Natural) {.noSideEffect.} =


genotrance:
Can we get a release for this?

c-blake:
Yeah.  I should fix up a few other places that need similar adjustment.  Give me a few minutes.

c-blake:
Ok.  New release published.  Let me know if you have any problems.  I was just running into these `csize` problems this morning myself.

c-blake:
I guess there's been some thrashing about on this. Maybe I should not have fulfilled @genotrance's request for a new release. In my defence, I think knowing what to do 15 days ago would have required clairvoyance about how Araq was going to handle things.  Anyway, I think I am going to just define `csize` as a module-internal type alias for `uint` in all the modules that use it.  Objections?  Affirmations?  Alternatives?  I just pushed a commit.  Speak up today before I re-release.

Clyybber:
I'm sorry, I couldn't have predicted that the change would be reverted (and tbh that revertion was pretty useless, as this change didn't cause much more breakage than what I've send PRs out for)..
RE: csize as an internal alias: I think thats a good idea.

c-blake:
Ok.  It's only a few lines and not drastically confusing.  Maybe a year from now, I'll drop support for old Nim versions. Then the story will be simpler.

I do think deprecation periods are nice, and Nim does have mechanisms for that, though types are trickier than overloads on procs. Transition care is always a bit of a judgement call.  Anyway, I'm just trying to do right by `cligen` users going forward not relitigate some Araq decision.  Just bore mentioning that I couldn't know the right thing to do.

PR #130, opened 2020-01-19
allow const variables to be used as doc; fixes #110

### LINKED ISSUE #110: initFromCL() doc and short parameters doesn't take values from variables.
Using nim v0.20.0 and cligen-0.9.31.

    const appsummary = "Description of this app"
    type App* = object
      output*: string     ## outputfile
      vargs*: seq[string] ## Arguments
    const appdef* = App(output: "nimcc")
    const appsummary = "Description of this app"
    const appshort = {"output": 'f'}
    
    when isMainModule:
      import cligen
      var app = initFromCL(appdef, doc = appsummary, short = appshort)
      echo "app: ", app

and then running it as: `myapp -h` doesn't display the content of string `Description of this app` string, but rather just the variable name `appsummary`

and `myapp -f example` doesn't work as `appshort` doesn't get applied to `short`

enthus1ast:
Signed-off-by: David Krause <david@code0.xyz>

i am by far an macro expert and this maybe has side effects that i'm unaware of.
But it seems to work.

```nim

import cligen
const help = """

  help
  ====
  my 
  help

"""

proc foo() = discard

dispatch(foo, doc = help)

```

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -267,7 +267,7 @@ macro dispatchGen*(pro: typed{nkSym}, cmdName: string="", doc: string="",
   let impl = pro.getImpl
   if impl == nil: error "getImpl(" & $pro & ") returned nil."
   let fpars = formalParams(impl, toIdSeq(suppress))
-  var cmtDoc: string = $doc
+  var cmtDoc: string = if doc.kind == nnkStrLit: $doc else: doc.getImpl.strVal
   if cmtDoc.len == 0:                   # allow caller to override commentDoc
     collectComments(cmtDoc, impl)
     cmtDoc = strip(cmtDoc)


enthus1ast:
maybe the same could apply for `shortH`

enthus1ast:
```nim
import cligen
const help = """

  help
  ====
  my 
  help

"""

proc foo() = discard
proc baa() = discard

dispatchMulti([foo, doc = help], [baa, doc = help])
```

c-blake:
The way you structured things to switch off the Nim node kind, I don't think side effects are likely.  So, I'm merging this for now.  We can revisit if it causes trouble and thanks for the patch!

I agree that the same approach would probably work for many more parameters which have simple Nim data types like `string`.  I always use literals in my own CL programs, but especially for large strings various compile-time string-builder helpers make a lot of sense.

Note that the user for https://github.com/c-blake/cligen/pull/110 requested that `short` work. `shortH` is just some internal abbreviation for `short["help"]`.  So, we should maybe keep that issue open.  Technically, what @0xACE specifically requested was invalid syntax.  One cannot just say `const name = { x: y }` in Nim.  `{}` is an incomplete/non-typed literal only syntax for what Lispers call "association lists".  It is not interpreted as an actual `Table` type (which is how it should be - there are several `Table`-like types anyway).  Rather `{}` is more like a `seq[key, val: untyped]` that must be compile-time converted to whatever is wanted *at the site where `{}` is used*.

Even so, `short` in `cligen` is notionally always really a `Table[string, char]`.  And `help` is always a `Table[string, string]`.  So, I think `short` and `help` could also be made to work with maybe a little more `cligen` work than you did for `doc`, and definitely with the user having to say something more like this than what he initially wrote:
```Nim
const appshort = {"output": 'f'}.toTable
```
I've actually had at least one version of Nim in the past year where that kind of `const Table` construction failed on me while picking apart the literal has never failed, but it has always been the intent of Nim core devs for such to work.

PR #143, opened 2020-05-05
Add precedence for config/config.toml lookup

kaushalmodi:
Add a useful error message if config.toml is found, but the project is
not compiled with parsetoml.

Ref: https://github.com/c-blake/cligen/pull/141#issuecomment-624030900

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -58,10 +58,7 @@ proc toInts*(x: seq[ClHelpCol]): seq[int] =
   ##Internal routine to convert help column enums to just ints for `alignTable`.
   for e in x: result.add(int(e))
 
-when defined(cgCfgToml):    # An include helpTmpl and syntaxHelp
-  include cligen/clCfgToml  # Trade parsetoml dependency for better documention
-else:
-  include cligen/clCfgInit  # Just use stdlib parsecfg
+include cligen/clCfgInit
 
 proc onCols*(c: ClCfg): seq[string] =
   ##Internal routine to map help table color specs to strings for `alignTable`.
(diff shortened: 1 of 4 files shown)


c-blake:
I'm not sure we want to do this.  I don't know how rigid in the TOML world in general always having `.toml` extensions are.  Also, `cligen` may be a leaf dependency of different packages...Some which compile with `-d:cgCfgToml` and some which do not.  Then a user might have a "stray" `config.toml` around from one but maybe doesn't use it for others, etc.  It's kind of a complicated situation.

c-blake:
We might even need `$CLIGEN` to have `$PATH` like search semantics with `:` separators even.

c-blake:
And if we ever do both parsers in one program we need to decide on/have the user decide on priority.

kaushalmodi:
>  I don't know how rigid in the TOML world in general always having .toml extensions are

I can say that it's pretty rigid and consistent. I have using the config.toml's in [Hugo ](https://gohugo.io/getting-started/configuration/#configuration-file), [Netlify](https://docs.netlify.com/configure-builds/file-based-configuration/) and [Gitlab Runner](https://docs.gitlab.com/runner/configuration/advanced-configuration.html).

> Then a user might have a "stray" config.toml around from one but maybe doesn't use it for others, etc. It's kind of a complicated situation.

Hmm. I agree.

> And if we ever do both parsers in one program we need to decide on/have the user decide on priority.

I think that's unnecessary complexity. If you choose just one precedence and document it, users can then choose which format config to keep/maintain. For example, I converted your example config to config.toml and will keep maintaining that.

So for more folks, the precedence wouldn't matter much as long as it's fixed and documented.

c-blake:
It may not be..exactly, CL user choice.  The CLI util author may turn on `-d:cgCfgToml`.  So, if a user justs `nimble install x` and `nimble install y` they may need both formats at the same time.  Then either mode warning about the other would be wrong.  But then, much of the time the CL user may be the one doing `-d:cgCfgToml`.  So, it's kind of tricky.  I think we should just leave it as-is now where neither parser knows about the other's existence.  At least for a while.

kaushalmodi:
OK, I will close this PR then.

PR #154, opened 2020-06-21
remove extranious closing parenthesis

AvdN:
The parenthesis were unbalanced, would not compile if cut and pasted


c-blake:
Oops.  Thanks for catching/fixing that!

PR #163, opened 2020-09-15
Support proc params as sym

cooldome:
Upcoming changes to Nim compiler will make proc params symbols instead of idents. Package tests showed cligen changes are required.

https://github.com/nim-lang/Nim/pull/15252

https://github.com/nim-lang/Nim/pull/15332


c-blake (approved):
Ok.  The PR looks backward & forward compatible to me.

c-blake:
All my tests compile fine with this, too, which you may already know.  Thanks for helping `cligen` stay current and have a good evening!

Issue #165, opened 2020-09-22
My fork has an error


pb-cdunn:
I don't even know how or why this test is running, but my fork fails on the `mast` branch, which is the `master` in your upstream repo. Any ideas?

* https://github.com/bio-nim/cligen/runs/1150792213

I see. The repo has [this](https://github.com/bio-nim/cligen/actions/runs/267333289/workflow):

```
.github/workflows/test.yml
```
The failure is for only `test (ubuntu-latest, stable)`; the other 3 tests pass. Are you seeing that too?

c-blake:
Yeah.  Personally, I just run `./test.sh` before I commit anything that could touch the basic functionality and think this git/CI stuff is over-rated and that is what I would recommend to you.   @jiro4989 and @kaushalmodi wanted/added this stuff.  I see some kind of failure whenever I `git push` anything.  If you have some idea how to prevent the failures, I'm open-minded.  When I look at those logs (like the one you sent), I only see observable stores warnings, not any real errors.  The failure only happens on ubuntu-latest because it only runs there, though.  If you look at the other 3 you will see there is no run of the test phase.

c-blake:
Well, I guess the "error" could simply be the non-zero exit of the `diff` program/script.  But that isn't really an "error".  The change in integer hash is known and the other stuff is just warning deltas.  Anyway, I don't think this is a real problem.

pb-cdunn:
Ok, I'll try to ignore github CI for now. This is my first experience with that. I guess it's a good thing that they provide it. I'm surprised that forks automatically run it though.

c-blake:
I am also slightly surprised forks auto-run it, but OTOH forks being basically "first class" makes it unsurprising.  I wouldn't be surprised if early on in GH history forks didn't run them.

Anyway, I could probably upgrade the Nim version in `test.yml` from 1.2.0 to 1.3.7 to maybe get an empty diff which might make the exit status 1 condition go away for now.

c-blake:
Well, I tried a bit to get the `diff` output down to an empty string/diff-exit 0 but failed.  I did get it to a more obviously unimportant looking delta, though, of just a bunch of "...." lines.  https://github.com/c-blake/cligen/actions/runs/269191877 .

[...]

PR #166, opened 2020-09-23
fix test script #165

jiro4989:
## Overview

- Version up 'setup-nim-action'
- Skip test on 'macOS' VM because devel-compiler doesn't exist
- Remove compiler-message from 'test/ref'
- Change test script

## Change test script

The current test script writes '--help' message and compiler message to 'test/out'.
It is difficult to ignore with 'grep'.
So I split 'compile' phase and 'run' command phase.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           path: ~/.nimble
           key: ${{ runner.os }}-nimble-${{ hashFiles('*.nimble') }}
-      - uses: jiro4989/setup-nim-action@v1.0.2
+      - uses: jiro4989/setup-nim-action@v1
         with:
           nim-version: 'devel --latest'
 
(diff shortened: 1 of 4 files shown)


jiro4989:
wait, I write description of this PR.

jiro4989:
I wrote description of PR.

jiro4989:
fmm, 'gh-pages' test failed.

c-blake:
It is good to keep those compiler messages in test/ref (or test/out) for when I change something or Nim changes something to produce new warnings/errors.  I'm not sure why the debian thing is spitting out "....." lines, but we could probably filter those out, too, like the `grep -v '^CC'`.

I'm not sure why I cannot silence that observable stores the ways I was trying.  Silencing breaks for me locally, too, but works fine if I do an `@if nimHasWarningObservable..` in my `$HOME/.config/nim.cfg` (that `@if` is actually used by compiler `nim.cfg`'s, too).

jiro4989:
fmm... I understood.

c-blake:
I merged this but then still had to fix some stuff to get it working.  Basically, I had some things in my `nim.cfg` altering my output slightly. Now they are on the cmd line in `test.sh`.  (I never really tried to match other-environment-build with my-environment-build before your CI stuff.)

TBH, I spent most of the time on this trying to get code-local warning deactivation of `[ObservableStores]` working.  It still does not and I still don't know why. :(

PR #171, opened 2020-10-24
Strip OSC codes

SolitudeSF:
also strip osc codes. required for https://github.com/c-blake/lc/pull/2
now the name of the procedure doesn't reflect its behavior exactly. should i rename?

### DIFF
--- a/cligen/textUt.nim
+++ b/cligen/textUt.nim
@@ -3,16 +3,27 @@ from terminal import terminalWidth
 from unicode  import nil
 import critbits, math, ./mslice # math.^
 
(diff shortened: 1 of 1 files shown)


c-blake:
I don't think you really need to rename.  If you did then `stripEsc` might be best for some generalized all escape sequence idea, but we would need some deprecated alias for `stripSGR` for a while.

c-blake:
I went ahead and de-deprecated that.  Probably no need to annoy people with the rename as most present uses probably don't have OSC codes to remove.

PR #188, opened 2021-02-26
Use Makefile rather than shell script to run tests

SirNickolas:
I must say, spawning 60 memory-hungry processes in parallel isn’t the best idea. My PC died when I tried to run `test.sh`. So I rewrote it with a Makefile.

Differences from the former script:

* Invoked with `make -j4` (or however many cores you have) instead of `./test.sh`.
* Additional arguments for the compiler are supplied with `make -j4 a='<args>'` syntax rather than `./test.sh <args>`.
* Diff’s output is colorized. You can undo it by removing `--color=auto` from line 31 if you’re not fond of colorful diffs.
* Bonus: `make clean` removes all build artifacts.

For the rest, it works exactly the same.


c-blake:
That looks like a GNUmakefile to me.  Migh help FreeBSD/OpenBSD folks to call it that.  I'll have to go over this more carefully in the morning.

c-blake:
A few more comments.

`gcc` uses a lot of memory/time.  If you find yourself running the test suite on a wimpy/small machine then `--cc:tcc` is much better.  That is what I personally use even on a big box.  E.g.:
```
$ BE='c --cc:gcc' cgmemtime ./test.sh
Child user:  550.641 s
Child sys :   28.751 s
Child wall:  146.768 s
Child high-water RSS                    : 157.4 MiB
Recursive and acc. high-water RSS+CACHE : 16482.8 MiB
```
vs.
```
$ BE='c --cc:tcc' cgmemtime ./test.sh
Child user:   71.371 s
Child sys :    9.096 s
Child wall:   20.544 s
Child high-water RSS                    : 158.2 MiB
Recursive and acc. high-water RSS+CACHE : 7602.6 MiB
```
I actually default to `--cc:tcc` in my `$HOME/.config/nim/nim.cfg`.  I use the [mob branch](http://repo.or.cz/w/tinycc.git) of tcc also mirrored at [github](https://github.com/TinyCC/TinyCC).

Really I usually run this on a big 16-core/32-corethread box with tcc in just 5 seconds.  Then I have some `nim.cfg` logic to let me say `-d:r` to switch both to `gcc` and to danger mode.  Used to be `-d:release` mode.  So, I do nothing for rapid dev/testing and add a `-d:r` for instead building a fast running program.

Anyway, I am not opposed to making this more controlled in its simultaneous resource usage, but it would be better to make it a `.nims` nimscript thing instead of (any kind of) a makefile.  If we are going to overhaul this, it would also be better to add behavioral tests beyond help generation.  (E.g., test `./foo --alpha 1.0  --alpha=1.0 -a1.0` all work the same, `./foo --alpha` errors out, etc.)

All this said, I am also not opposed to just including a GNUmakefile "on the side" for now.

SirNickolas:
Sorry about deleting `test.sh`, forgot about the CI.

Thank you for pointing out `tcc`! I used to use it for debug builds but *temporarily* disabled it in my config some time ago. With it, execution time (using `make`) dropped from 211 seconds to just 68 on my machine.

However, when compiling with TCC, I get a SIGSEGV for `FullyAutoMultiTopLvl.out`. Will look into it later.

```
Traceback (most recent call last)
cligen/cligen.nim(1041) FullyAutoMulti
cligen/cligen.nim(720) multiSubs
cligen/cligen.nim(701) parser
cligen/cligen/parseopt3.nim(180) initOptParser
nim-1.4.4/lib/system/assign.nim(177) genericSeqAssign
nim-1.4.4/lib/system/assign.nim(141) genericAssign
nim-1.4.4/lib/system/assign.nim(100) genericAssignAux
nim-1.4.4/lib/system/assign.nim(72) genericAssignAux
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

By the way, you can write `DIFF ?= diff` in the makefile so that you don’t need `-e` switch for `gmake`.

I tried to mock with NimScript; however, its parallelization abilities leave much to be desired. Importing `std/asyncdispatch` errors out with `cannot evaluate at compile time: currentID`, and both `parallel` and `spawn` are undeclared identifiers. `osproc` also cannot be imported: `undeclared identifier: 'alloc0'`—so we’re not able to write our own event loop even if we wanted.

I guess, we should patch the makefile so that it is understood by more dialects of Make. Or switch to another build system.

c-blake:
I just pushed the `?=` for you.  If you prefer a different variable name from `DIFF` (e.g. `CLIGEN_DIFF`) then now is the time to speak up. :-)  Pros & cons to same as other use cases.

I think `gmake` is almost everywhere, and it's only a build system to run tests.  So, I don't know how much more effort makes sense.  I have a little parallelizer program I call `stripe` I haven't distributed that I could add to the `examples/` and `cligen` devs could then use that in `test.sh`.    

Anyway make -j4 or whatever works now, and I really do think gmake is common enough to not worry.  I just think it should be specifically advertised as a `GNUmakefile`.  It might even be unusual enough that people go `head GNUmakefile` and read something useful.  ;-)

I also tend to think more value add would come from behavioral testing or even coalescing some distinct `test/` programs that could maybe be just one program with a `when` fork like `test/FancyRepeats*.nim`.

c-blake:
Oh, and I just had success with a fresh git clone of that `tcc` mob branch and nim version 1.2, 1.4, and 6278b5d89af38e90aa30cfc4c217a2f4b1323339, but it is absolutely an under-supported mode of operation as came up most recently in [The Forum](https://forum.nim-lang.org/t/7337).

I don't know if/when incremental compilation (IC) will be ready, but that should speed things up a lot as well, but will also probably have trouble working with parallelism.  At one point, `tcc` was maybe going to be the basis for a better REPL than `nim secret`, but  then INim arose and interest was too low or something.

c-blake:
Oh, wait...I was using tcc fb1fb8219ccf4813d63edf8c40fddc756e0c60cc.  Sorry.  Let me check with a fresh git clone.

c-blake:
Well, also works for me with tcc d0d0c8b688366ca9dccddcfbacb5e33c8b1790bb which is from last Sunday not last April.

SirNickolas:
I have TCC v0.9.27 (released 3 years ago); I think it’s a bug somewhere in the codegen. I’ll try installing from Git and tell if that still happens.

> If you prefer a different variable name from `DIFF` (e.g. `CLIGEN_DIFF`) then now is the time to speak up. :-)

No, the name is fine. But you left `-e` in the comment.

c-blake:
Fixed the -e.  Beware that `tcc -v` for all the git releases still prints that "v0.9.27".  But if your file time stamp on `.../bin/tcc` says 3 years ago then, yeah, you should upgrade.

SirNickolas:
Installed d0d0c8b688; still reproducing. TCC passes its own test suite—it’s Nim-generated code that is causing problems. Investigating.

c-blake:
Gotcha.  That's the current HEAD.

Hmm.  I do have a custom 100 line patch that I drop in my `/etc/portage/patches/dev-lang/tcc` to colorize warnings & errors without passing through a text filter (which would slow things down..a lot).  It should not impact compiling, though.

Nim generated code often has trouble, though, on one or another backends like js/c++/tcc/etc.  What version of Nim are you using? It sounds like you fetched a bunch to help me test.  So, you could try a few to isolate when the problem started/went away.

c-blake:
Well, I emailed you that patch and my ~/.config/nim/nim.cfg line, just in case.

SirNickolas:
So, what I’ve been doing and have figured out so far.

First of all, even if you compile with `--cc=tcc --debuginfo=on`, GDB will fail to extract debug information from the binary and the only option will be debugging assembly with no symbolic names. I decided not to dive into it and look for other ways.

`tcc`’s help revealed `-b` switch: “compile with built-in memory and bounds checker (implies -g).” Out of curiosity, I tried it (`--cc=tcc --debuginfo=on -t=-b -l=-b`), and for some reason that time GDB was able to read debug info. However, it quickly became obvious that TCC and GDB don’t understand each other well:

```io
(gdb) pt NI
type = <range type>
(gdb) # hmm... what?
(gdb) whatis NI
type = long
(gdb) p sizeof(long)
$1 = 8
(gdb) p sizeof(NI)
$2 = 4
(gdb) # wtf?
```

But it is not related to the segfault, and just makes debugging slightly more cumbersome. Moving on.

I ran the program, caught SIGSEGV, and found myself in the middle of `markStackAndRegisters` phase of the GC. In [this template][root-marking], to be exact. It dumps registers to the stack, marks them, and continues digging deeper through the stack until it reaches the bottom. Obviously, this doesn’t work because of TCC’s memory checker so I put calls to `__bounds_checking` to temporarily disable it while scanning was in progress.

After re-running the program, I received a SEGV in a [different place][set-length] because `extGetCellType`, which reads some field from some GC-related structure, returned `nil`. Luckily, this call is placed in a conditional compilation branch so I didn’t bother to investigate further and just switched to another GC backend.

I recompiled with `--cc=tcc --gc=markAndSweep --debuginfo=on -t=-b -l=-b` and ran, and it didn’t crash, hooray! Instead, it printed usage *twice*, then entered an infinite loop. “Excellent,” I thought, “it works twice as much as required and doesn’t even want to stop.”

After following the execution in the debugger and reading some disassembly, I found the problem. For this C fragment (from `cligenQuit` inlined into `cligenScope`):

```c
TM__24FiCw3PcCVXKwDFJfTvrg_3.status = setjmp(TM__24FiCw3PcCVXKwDFJfTvrg_3.context);
if (TM__24FiCw3PcCVXKwDFJfTvrg_3.status == 0) {
```

TCC generated the following machine code:

1. Invoke `setjmp`.
2. Spill its return value to memory location *A*.
3. Copy it to memory location *B*. (Okay, `-O0` is allowed to do dumb things.)
4. Reload it from memory location *C*. (???)
5. Compare it to 0.

(There are several calls to `__bound_ptr_indir8`—that is the address-sanitizing function—in between the steps, which I omitted for clarity.) So there is really a bug in the compiler’s codegen. Sweet.

At last, I decided to try `--cc=tcc --gc=markAndSweep` (without any debugging options). Surprisingly, it did work correctly. Moreover, all `cligen`’s tests were passed successfully with these flags.

Summing up:

1. GDB is unable to read debug information from Nim programs (even `echo "Hello World"`) with `tcc` backend. `--debugger=gdb` doesn’t help. On the other hand, it has no problems with a C program compiled with `tcc -g`. Could you check on your machine, please? Do you have any idea what we can do with it?
2. TCC generates incorrect code sometimes if you pass it `-b` switch.
3. Your program might crash if you use `refc` GC (default). `arc`, `orc`, `markAndSweep`, `boehm`, `regions`, and `none` work fine (didn’t check `go`). Could either be a bug in Nim’s runtime, or TCC interpreting it incorrectly. Cannot tell more since I’m not able to debug it.

[root-marking]: https://github.com/nim-lang/Nim/blob/v1.4.4/lib/system/gc_common.nim#L376
[set-length]: https://github.com/nim-lang/Nim/blob/v1.4.4/lib/system/sysstr.nim#L355

Nim 1.4.4, TinyCC d0d0c8b688, GDB 10.1.

SirNickolas:
Please, ignore item 1. For debug info to be available, you have to pass `-g` to both tcc-compiler and -linker, but Nim’s `--debuginfo=on` only affects compilation. Adding `-l=-g` helps.

c-blake:
Oops!  I neglected to check/mention that my `nim.cfg` defaults to `gc = markAndSweep`!

That induced me to do other experimental compiler option tweaking & test running.  I began running into infinite loop problems (in `test/MultiMulti`, not `test/FullyAutoMult`) when I was fiddling with `--tlsEmulation:on` (which I usually have turned on for `tcc` and turn of for `gcc`).  I have recently been defaulting to `--threads:on` and for `tcc` that requires the TLS flag.  (The compiler helpfully barks at you in this instance.)

I cannot quite reproduce all your reported behaviors, but an infinite loop in a `dispatchMulti` setting is a lot closer than before.  Maybe I should email you my whole `nim.cfg` file.

c-blake:
I just emailed you my `nim.cfg`.  I also have trouble with the default garbage collector, though for `test/MultiMulti` infinite looping (and not at the top level) not `test/FullyAutoMulti`.

c-blake:
(I should say that I am still using test.sh and your `gmake`-based approach which you are probably using could run tests in a different order.)

c-blake:
Oh, and you should probably confirm that I am emailing to the right address. (Sure seemed plausible, but best to ask.)

SirNickolas:
Thank you, I’ll continue investigating it. Special thanks for the coloring patch. (Didn’t you consider merging it to upstream?)

> (I should say that I am still using test.sh and your gmake-based approach which you are probably using could run tests in a different order.)

It does. If you’d like it to be deterministic (as much as it’s possible with multiprocessing), move `$(sort )` from line 49 to line 46 (`$(sort $(TESTS_OUT) $(TESTS_TOP_LVL_OUT))`).

c-blake:
I guess when I wrote that like 10+ years ago I was not enough of a "mobster" to put much thought into pushing it upstream to the mob branch. ;-)  I probably should, though.  Or if I get hit by a bus you can.

Truth is that I think I've only had to re-org the patch from upstream code evolution like 2..3 times. Just leaving it in my /etc/portage/dev-lang/tcc directory is super easy.  So, laziness has won out on this one.  In point of fact, one of the few re-orgs was just the other day.  So, it may have a bug. :)

I have a lot of undistributed code, including a window manager, web server, various libraries, piles of command-line utilities, etc.  It often takes a social component (someone talking about it with me) to get me to publish.

Order of testing-wise, `test.sh` is pretty unordered - *except* the two mentioned programs being before the huge shallow forest.  So, it just seemed worth mentioning since I think we are pretty close to both seeing the same problems.

c-blake:
Oh, if you get to the point of doing a Nim git bisect (extreme, I know -- even just spot checking), I think I only added `--threads:on` and `--tlsEmulation:on` by default at the very end of this January - 4 weeks ago, but I have always defaulted to `tcc` for `./test.sh`.  So, my guess would be that older Nim's will at least not fail in the same way and recent code is the culprit.  Just a guess.

PR #196, opened 2021-05-24
Update default `hTabSuppress` value in docs

ZoomRmc:
The docs still had the superseded "SUPPRESS" value for `hTabSuppress`.
Also expanded wording a bit.


c-blake:
Thanks.

PR #210, opened 2022-06-07
allow different file descriptors & input / output object sizes

Vindaar:
This is just a pretty straightforward extension to assign custom file descriptors to use in the kid ⇔ parent communication. The user currently of course has to make sure those exist and can be used. Ideally we would be able to take care of that for the user and provide a `stdreq` / `stdrep` directly in the `work` procedure. 
That could be done either by handing a `stdreq` and `stdrep` argument to `work` or by defining some template of that name?

In addition I replaced `aux` by separate `auxIn` and `auxOut`. But `auxIn` as far as I can tell is actually not even needed (at least not currently). So possibly `aux` can just remain a lone argument always referring to the size of the object for `framesOb`. 

*edit*: Also added a `read` proc for `MSlice` to read into a `var T` object using `copyMem`. I took used `read` instead of `uRd`, because the u = unlocked meaning for `File` is unrelated here. Oh, and we might want to turn that `doAssert` into an `assert` of course.

Here's some example code that shows how to use separate streams with this now:

```nim
import cligen / [procpool, mslice, osUt], cligen, std / [sequtils, os, streams]

type
  ProcData = object # data we hand to each kid worker process
    id: int
    σs: float
    σb: float
    nmc: int

  ProcResult = object # data that is returned as a result
    id: int
    res: float

proc main(jobs: int) =
  var stdreq: File = open("/dev/tty", fmRead) # open our custom streams using `/dev/tty`
  var stdrep: File = open("/dev/tty", fmWrite)
  flushFile stdreq # not needed?
  var pp = initProcPool(
    (proc() =
       var p: ProcData
       while stdreq.uRd(p): # read `ProcData` sized data from our custom input
         echo "hello from one worker! with line: ", p # can output to `stdout` just fine
         var cnt = 0 # do "work"
         for _ in 0 ..< 1_000_000: cnt += 1
         echo "done ", cnt # more output!
         var pRes = ProcResult(id: p.id, res: cnt.float)
         doAssert stdrep.uWr(pRes) # write back to our custom output stream
         flushFile stdrep # needed?
    ),
    # different input and output sizes
    framesOb, jobs = jobs, auxIn = ProcData.sizeof, auxOut = ProcResult.sizeof,
    # and our custom file descriptors
    stdreq = stdreq.getFileHandle(), stdrep = stdrep.getFileHandle())

  # define some data
  let ds = @[ProcData(id: 0, σs: 0.1, σb: 0.1, nmc: 1000),
             ProcData(id: 1, σs: 0.1, σb: 0.2, nmc: 1000),
             ProcData(id: 2, σs: 0.1, σb: 0.3, nmc: 1000),
             ProcData(id: 3, σs: 0.2, σb: 0.1, nmc: 1000),
             ProcData(id: 4, σs: 0.2, σb: 0.2, nmc: 1000),
             ProcData(id: 5, σs: 0.2, σb: 0.3, nmc: 1000)]
  var results = newSeq[float](ds.len)
  var getRes = proc(s: MSlice) =
    var p: ProcResult
    s.read(p)
    doAssert p.id < results.len, "Parsed bad data"
    results[p.id] = p.res
  pp.evalOb ds, getRes
  echo results # got data back correctly despite stdout usage!

when isMainModule:
  dispatch main
```

### DIFF
--- a/cligen/mslice.nim
+++ b/cligen/mslice.nim
@@ -111,6 +111,12 @@ proc mrite*(f: File, mses: varargs[MSlice]) {.inline.} =
   ## unlocked write all ``mses`` to file ``f``; Be careful of many fwrite()s.
   for ms in items(mses): f.urite ms
 
+proc read*[T](m: MSlice, ob: var T) =
+  ## Assumes `m` starts at beginning of object `ob`! Reads data in slice
+  ## into object `ob` using `copyMem`.
+  doAssert m.len >= ob.sizeof
+  copyMem(m.mem, ob.addr, ob.sizeof)
+
 proc `==`*(a: string, ms: MSlice): bool {.inline.} =
   a.len == ms.len and cmemcmp(unsafeAddr a[0], ms.mem, a.len.csize) == 0
 proc `==`*(ms: MSlice, b: string): bool {.inline.} = b == ms
--- a/cligen/procpool.nim
+++ b/cligen/procpool.nim
@@ -66,7 +66,7 @@ proc len*(pp: ProcPool): int {.inline.} = pp.kids.len
 
 proc close*(pp: ProcPool, kid: int) = discard pp.kids[kid].fd0.close
 
-proc initFilter(work: proc(), aux: int): Filter {.inline.} =
+proc initFilter(work: proc(), aux: int, fd0, fd1: cint): Filter {.inline.} =
   result.aux = aux
   var fds0, fds1: array[2, cint]
   discard fds0.pipe         # pipe for data flowing from parent -> kid
@@ -74,8 +74,8 @@ proc initFilter(work: proc(), aux: int): Filter {.inline.} =
   case (let pid = fork(); pid):
   of -1: result.pid = -1    #NOTE: A when(Windows) PR with CreatePipe,
   of 0:                     #      CreateProcess is very welcome.
-    discard dup2(fds0[0], 0)
-    discard dup2(fds1[1], 1)
+    discard dup2(fds0[0], fd0)
+    discard dup2(fds1[1], fd1)
     discard close(fds0[0])
     discard close(fds0[1])
     discard close(fds1[0])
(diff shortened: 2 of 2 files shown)


c-blake:
So, I went a slightly different way here (but applied your `mslice.nim` patch separately).  Your new main should look something like this:
```Nim
proc main(jobs: int) =
  var pp = initProcPool(
    (proc(r, w: cint) =
       var p: ProcData
       let i = r.open
       let o = w.open(fmWrite)
       while i.uRd(p): # read `ProcData` sized data from our custom input
         echo "hello from one worker! with line: ", p # can output to `stdout` just fine
         var cnt = 0 # do "work"
         for _ in 0 ..< 1_000_000: cnt += 1
         echo "done ", cnt # more output!
         var pRes = ProcResult(id: p.id, res: cnt.float)
         doAssert o.uWr(pRes) # write back to our custom output stream
    ), framesOb, jobs = jobs)
```
At least with a fresh git pull of `cligen` this compiles and runs producing:
```
hello from one worker! with line: (id: 0, σs: 0.1, σb: 0.1, nmc: 1000)
hello from one worker! with line: (id: 1, σs: 0.1, σb: 0.2, nmc: 1000)
done 1000000
hello from one worker! with line: (id: 2, σs: 0.1, σb: 0.3, nmc: 1000)
done 1000000
hello from one worker! with line: (id: 4, σs: 0.2, σb: 0.2, nmc: 1000)
done 1000000
hello from one worker! with line: (id: 3, σs: 0.2, σb: 0.1, nmc: 1000)
done 1000000
hello from one worker! with line: (id: 5, σs: 0.2, σb: 0.3, nmc: 1000)
done 1000000
done 1000000
@[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

Vindaar:
Closing this as it's now redundant. :+1:

c-blake:
Yeppers - I was going to if you didn't. :-)

PR #214, opened 2022-07-05
add `nextSlice` helper to advance slice to sep and store in a given mutable slice

Vindaar:
This is similar to the `readLine` procedure for the `Stream` type in
the stdlib, except working on slices.

The implementation is obviously directly based on the `mSlices` iterator. Initially I wanted to base one on the other, but I decided against it for now. This way I can add a `nil` check in `nextSlice` and don't need to shadow the input `mslc` argument in `mSlices` to have a slice that we advance for each yielded slice.


c-blake:
Looks good to me.  Feel free to re-PR if you discover any issues.

PR #215, opened 2022-07-05
fix broken docs due to nim doc rst / markdown changes

Vindaar:
Should hopefully fix the CI. At least it makes the docs generate successfully again locally.


c-blake:
Thanks!

PR #217, opened 2022-07-24
replace shallowCopy with move

ringabout:
ref https://github.com/nim-lang/Nim/pull/20070

Imo, `shallowCopy` only works for string/seq and is about to be removed for ARC/ORC. Inn this case, `move` is better.


c-blake:
Ok.  Thanks!

PR #221, opened 2023-02-24
explicit convert bool types to int types for quoted variables

ringabout:
Here is the definition of `parReq`, which is an int type
```nim
parReq*: int        ## flag indicating parameter is mandatory
```

Here is the definition of `isReq`, which is a boolean type

```nim
let isReq = if i in mandatory: true else: false
```

It is not safe to rely on the type erasure at the compile time, which might be changed in the future => https://github.com/nim-lang/Nim/pull/21433


c-blake:
Thank you!!

PR #227, opened 2023-07-13
Enable most warnings for tests

SirNickolas:
So I did what we discussed in #225. I put `GcUnsafe` and `GlobalVar` in `nim.cfg` since cligen uses global variables; there’s nothing a person who wishes to run tests can do with it. `Deprecated` remained in `GNUMakefile`/`test.sh` so you can easily run without it via `w= ./test.sh`; and so did `ProveField` (arguable; might be better to move it to `nim.cfg` too).

As to a strange cat in the pipeline, I suspect you tried `2>&1 >$o`, and it didn’t work because the order of redirections is important. They are processed from left to right: first, stderr is sent to where stdout is _currently_ connected to; then, stdout is redirected to a file. And we want this to happen vice versa. (I ran `test.sh` without final `&`, and the tests passed.)

`hint[Performance]` is indeed issued at a later stage and cannot be scoped, as you told.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -23,8 +23,11 @@ NIM := $(or $(nim),nim)
 NIM_BACKEND := $(or $(BE),c)
 
 NIM_FLAGS := \
-	--hints:off --warning:all:off --warning:User:on --colors:off\
-	--parallelBuild:$J $(NIM_EXTRA)
+	--verbosity:2 --warning:Deprecated:off --warning:ProveField:off \
+	--hint:Path:off --hint:Conf:off --hint:Processing:off --hint:CC:off \
+	--hint:Exec:off --hint:Source:off --hint:Link:off --hint:SuccessX:off \
+	--hint:GCStats:off \
+	--colors:off --parallelBuild:$J $(NIM_EXTRA)
 NIM_CACHE := $(HOME)/.cache/nim
 
 TESTS_OUT := $(patsubst %.nim,%.out,$(wildcard test/[A-Z]*.nim))
--- a/test.sh
+++ b/test.sh
@@ -2,17 +2,15 @@
 exec < /dev/null
 export COLUMNS="80" CLIGEN_WIDTH="80" CLIGEN=/dev/null
 rm -rf $HOME/.cache/nim/*
-v="--verbosity:1"
-h="--hint[Processing]=off --hint[CC]=off --hint[Exec]=off --hint[SuccessX]=off"
-: ${w="--warning[ObservableStores]:off --warning[Deprecated]:off --warning[ImplicitDefaultValue]:off"}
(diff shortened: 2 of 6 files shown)


c-blake:
This looks pretty good to me, but I'd like to run it against my set of old Nim's before merging.  That may have to wait until tomorrow.

c-blake:
I suspect that `--verbosity:2` causes some trouble with older Nim's (< Nim-1.4) printing out the C compile lines (which induces big `diff`'s).  I doubt anyone but me runs the test suite on such old Nims, but I may change that back in a quick follow-up commit...

SirNickolas:
Strange. `--hint[CC]:off --hint[Exec]:off` should have disabled that.

c-blake:
I was not very clear.. for me it was *not* printing out of `gmake`, but it was printing the CC line out of `test.sh`..

SirNickolas:
Yeah, just figured that out.

SirNickolas:
Another mystery: old Nims emit those compilation lines even on `gmake` — if you tell it to build a specific target (say, `gmake J=0 test/AllSeqTypes.out`). Seems like Nim did some non-obvious check. No idea what it can be (adding/removing `< /dev/null` and `| cat` doesn’t help).

For now, the best I can think of is adding

```nim
@if nimHasInvariant: # 1.2+
  verbosity = 2
@end
```

and removing that switch from `GNUmakefile` and `test.sh`.

c-blake:
Yeah.  Or we could have it be `verbosity:1` in the `test.sh` & `GNUmakefile` and then conditionally bump it up to `2` in the config.  I'm fine with either.  (I almost never run it changing `$v` anyway.)  I toyed with the idea of just moving only `test.sh` back to `verbosity:1`, but probably best to keep it consistent with the `GNUmakefile`.  Happy to merge any reasonable PR along any of these lines.

(BTW, don't worry - I am still investigating your `tmplParse2` as to whether the perf really need to change the API and not just the impl, though I do truly doubt anyone but me uses the API.  I only got full electricity restored last night.)

SirNickolas:
> Or we could have it be `verbosity:1` in the `test.sh` & `GNUmakefile` and then conditionally bump it up to `2` in the config.

Not exactly, `--verbosity` on the command line overrides one in the config. To conditionally bump it, we have to not pass it explicitly.

(You are doing great job. It’s the fastest support here among all the open-source projects I’ve encountered, and I’m not pressed for time on this issue.)

c-blake:
Well, for now I just did https://github.com/c-blake/cligen/commit/07b7a8d4fa8a2b8ed569e3d30cac5eb95b3c60cf as a kind of ham-fisted override to go back to `--verbosity:1` for those older Nim's.  Definitely let me know if you ever figure this out.

PR #228, opened 2023-08-25
Fix `parseopt3.getopt`  usage example

ZoomRmc:
A tiny doc fix to process the `CmdLineKind` enum exhaustively.

Perhaps a candidate for a runnableExample (yet to be used in cligen)?

### DIFF
--- a/cligen/parseopt3.nim
+++ b/cligen/parseopt3.nim
@@ -317,6 +317,7 @@ iterator getopt*(p: var OptParser): GetoptResult =
   ## Example:
   ##
   ## .. code-block:: nim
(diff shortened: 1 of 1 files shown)


c-blake:
Yeah... I never added `runnableExample` s.  We can PR those, too if you like.  Happy to just merge this if it's ready.

ZoomRmc:
> Yeah... I never added `runnableExample` s. We can PR those, too if you like. Happy to just merge this if it's ready.

Well, adding them will require a considerable extension of the provided examples to make them fully compile. If you're ok with it, I'll add one here (and possibly to some other user-facing routines).

PR #233, opened 2024-08-28
fix generic instantiations of `doArgParse`

metagn:
`doArgParse` takes 2 generic parameters, but in `argParse` for each number type, only 1 is given, which compiles due to bugs in Nim. To safeguard against the Nim compiler, the missing generic parameter is now provided.


c-blake:
Looks good to me.  Thanks!

It's pretty easy to forget a parameter.  Hopefully newer Nim versions will at least warn?

metagn:
> Hopefully newer Nim versions will at least warn?

The PR to Nim I'm working on, https://github.com/nim-lang/Nim/pull/24010, makes them error, which is how I found this. I'm considering the idea of allowing something like `doArgParse[BiggestInt, auto]` for cases like this where the compiler can infer the generic param so the user doesn't have to think about the value.

PR #235, opened 2024-10-27
Introduce "minStrQuoting" option to only quote string defaults when necessary

AngelEzquerra:
When this option is enabled, string defaults will not be quoted if they are non empty and do not contain any characters that would require quoting them when using the command line (e.g. when they do not contain spaces, new lines, globs, etc.)

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -46,6 +46,7 @@ type    # Main defns CLI authors need be aware of (besides top-level API calls)
     render*:      proc(s: string): string ## string->string help transformer
     widthEnv*:    string         ## name of environment var for width override
     sigPIPE*:     ClSIGPIPE      ## `dispatch` use allows end-user SIGPIPE ctrl
+    minStrQuoting*: bool         ## Only quote string defaults when necessary
 
   HelpOnly*    = object of CatchableError ## Ok Ctl Flow Only For --help
   VersionOnly* = object of CatchableError ## Ok Ctl Flow Only For --version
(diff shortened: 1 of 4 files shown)


c-blake:
This looks fine to me.  If we want end CLusers to be able to override CLauthor defaults then we should also add a little parse of that `minStrQuoting` to `cligen/clCfgInit.nim` and `cligen/clCfgToml.nim` for TOML users like @kaushalmodi.  That can be done in this or a separate PR, whichever is easier.  Just let me know.

kaushalmodi:
I don't see TOML spec mention that quotes are optional: https://toml.io/en/v1.0.0#string

c-blake:
Oh, that was maybe unclear.  The TOML is unchanged (except to parse the new option).  The option is about the generated help message where @AngelEzquerra wants to drop the quotes "most" of the time as a new non-default behavior.  Whether one's shell allows for quoting/whatever is, of course, shell-dependent.  The `{' ', '\t', '\n', '*', '?', '$'}` in this PR might become replaced by a CLuser-selectable charset at some point.

kaushalmodi:
Ah OK. Thanks for clarifying. That's not an issue then.

c-blake:
And since the default will be `false`/`off`, only new users/new programs should care/notice.  Any old TOML files should still work just not mentioning the new flag.

I did mention to @AngelEzquerra that he *could* just have his own `include cg/AngelE` that he used in all his progs, but as should be clear by now, I'm also pretty accepting of simplifying user's lives if that doesn't impact other users { like distributing `cligen/clCfgToml.nim` / `-d=cgCfgToml` itself, even though it means updates to both whenever we add anything.  [ But that's ok, too!! and better in one place than N if N users compile for TOML configs.. beats me how many do, and sure, there's probably some way to abstract over both if we really wanted, but 2 isn't very many. :-) ] }  Anyway, sorry to worry you with brevity.

AngelEzquerra:
I just pushed a new commit that adds support for this new option to the [layout] section of the config files. Hopefully this works as expected.

c-blake:
Looks good to me.

PR #237, opened 2024-11-02
Make it possible to use the raw, unwrapped docstring

AngelEzquerra:
This adds two ways to make cligen show the "raw" docstring, avoiding the automatic wrapping that is usually applied to it before display. The default behavior remains unchanged.

The first way to enable this "raw" behavior is by setting the new `rawDoc` configuration flag in `clCfg`.

The second way is by using the new `rawdoc` template variable in the `usage` template, instead of the existing `doc` variable.

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -49,6 +49,8 @@ type    # Main defns CLI authors need be aware of (besides top-level API calls)
     minStrQuoting*: bool         ## Only quote string defaults when necessary
     trueDefault*: string         ## How to render a default value of "true"
     falseDefault*: string        ## How to render a default value of "false"
+    noWrapDoc*: bool             ## Disable the automatic wrapping of the docs
+    noWrapTable*: bool           ## Disable the automatic wrapping of tables
 
   HelpOnly*    = object of CatchableError ## Ok Ctl Flow Only For --help
   VersionOnly* = object of CatchableError ## Ok Ctl Flow Only For --version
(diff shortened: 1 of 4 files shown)


c-blake:
A few things:

 - the adjective "raw" here is ambiguous because at least in this initial PR, things like \*foo\* or backticks do get colorized ("rendered" in the terms of the current code and as is common parlance in the HTML world).  So, in the cooking metaphor the text is "partly cooked", and further this does change the length of the text if you are doing any kind of manual text alignment across multiple rows.  I would suggest "nowrap" as a name for this to disambiguate.  This is a somewhat minor point, though.

 - word wrapping is currently done in 4 different places, not only the `doc` part.  The other places I can think of are:
    - the proc doc addressed here
    - each individual `help[param]` text in the main options table (via `alignTable`)
    - top-level text for `dispatchMulti`
    - subcommand list for `dispatchMulti` (also via `textUt.alignTable`)
    
So, what we have is that the question is raised as to what is better:
  - one global word wrap disable (perhaps at the 3 call sites called)   OR
  - 3 or 4 separate switches for activating wrapping, such as a new `ClCfg.wrap: set[AreaEnum]`
  
Relevant to why I think the 2nd option is better is that in a `dispatchMulti` setting, `PassValuesMulti.nim` shows how each subcommand can have its own `ClCfg`.  So, if the CLauthor is not even doing config files, this gives them total control over all the places `wrap` is called.  There are still some "loops", though (such as over all the options & all the subcommands) where I could still imagine someone very picky about the look of the generated documentation wanting wrap control just for some parameters or subcommands but not others.

It's also possible that instead of a boolean flag such as `area in clCfg.wrap`, it would be better to actually specify 2 or 4 different "wrap widths" (in units of terminal columns not counting colors, i.e. as measured by `cligen/textUt.printedLen`), where there was just some conventional width like `-1` that meant "never wrap".  (It could be 2 not 4 since the specially named "multi" sub-command can kind of be the 2 for the top-level/outer doc and the subcommand table while the 2 for the doc & options table could be on each subcommand in a dispatchMulti.

So, sorry for the length - I'm really just trying to talk out the problem to converge to a good long-term solution here.  What all these considerations have me thinking that might be is more like: `ClCfg` gets 2 new integer fields - `wrapDoc`, `wrapTable` with special values `-1`=never, `0`=run-time terminal width, `> 0`=wrap at that width. { This even leaves open a possible `-2` (< 0 by more than 1) to mean something else, such as "wrap at that many multiples of the terminal width" or something to bound how crazy long lines can get. }  This is admittedly a *little* bit "`-1` & `0` need explaining to the user - somewhere", but it feels like it has the right amount of flexibility.

Looking back over the github history in Issues with a search of just "wrap" turns up an awful lot of interest in this.  @Vindaar @kaushalmodi and @SolitudeSF have all had thoughts at various points in `cligen` history, as have @kraptor @halloleo @vitreo12 and probably a few others I'm neglecting.  But I'm tagging them all due to the broad interest in the topic and its potential resolution.

c-blake:
https://github.com/c-blake/cligen/pull/238 is a concrete stab at this.  (You can add ".patch" to the end of the URI to get a patch and apply it to your local `cligen` checkout if you want to play around with it.)

As a CLauthor you can just set `clCfg.doc = -1` in the 238 PR to get the effect of this 237 PR, but this also handles tables in the two `alignTable()` calls (which also call `wrap`, as mentioned).

AngelEzquerra:
After reviewing PR #238 and discussing it with @c-blake, we agreed that we don't need this PR anymore.

PR #239, opened 2024-11-03
Rename falseDefault and trueDefault to falseDefaultStr and trueDefaultStr

AngelEzquerra:
This will make it easier to understand what these options do.


c-blake:
This is fine, but you should also edit `RELEASE_NOTES.md`.

PR #240, opened 2024-11-06
Add $usehdr template variable

AngelEzquerra:
Also avoid adding the usage header at the top of the help message when the usage template string is customized and the $usehdr template variable (or the _default_ "Usage:\n" header) is used in it.

Note that this does not change the default behavior in any way. It only makes it easier to completely customize the help message by simply changing the usage template string without having to also set the `noHdr` flag.

### DIFF
--- a/cligen.nim
+++ b/cligen.nim
@@ -606,9 +606,11 @@ macro dispatchGen*(pro: typed{nkSym}, cmdName: string="", doc: string="",
                 else:
                   (if `cf`.useHdr.len > 0: `cf`.useHdr else: clUseHdr) &
                     (if `usageId`.len > 0: `usageId` else: `cf`.use)
+      let useHdr = if `cf`.useHdr.len > 0: `cf`.useHdr else: clUseHdr
       let argStart = "[" & (if `mandatory`.len>0: `apId`.val4req&"," else: "") &
                      "optional-params]"
       `apId`.help = use % ["doc",     hl("doc", indentDoc),
+                           "usehdr", hl("usehdr", useHdr),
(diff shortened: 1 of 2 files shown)


c-blake:
This seems a good start!

One logic thing I see is that `std/strutils.%` can accept either `"$usehdr"` or `"${usehdr}"` substrings.  So, you probably want that `"$usehdr" notin use` to become `not("$usehdr" in use or "${usehdr}" in use)`. No?

I'd say it's fine to alter the location of `$usehdr` to before `${doc}` in the `usage` template in the altered `test/PassValues.nim` (which notably does ***not*** `include cligen/mergeCfgEnv` and so will not pick up some global CLuser setting).  Related to the prior point, you don't *need* braces after `$` there since unlike `$doc` it isn't followed by literal `"Options:\n"` text, but they are allowed & I could see someone's impulse being to use them.[^1]

Having updated `test/PassValues --help` output, we'll want to fix the reference output (I usually just `cp test/out test/ref` after manual inspection of `./test.sh` output.).  Then finally, we'll want a `RELEASE_NOTES.md` entry for this for the upcoming version 1.7.8.  I can do these last two bits post-PR-merge if you want, though.

[^1]:  I often use these things as demo/URI link targets for answers to questions (so the asker gets some "fully worked out context" example).  A big FAQ/page of such pointers is one way we could expand the Wiki to actually document the 50..100 features/HOWTOs probably documented mostly in `RELEASE_NOTES.md` or source/config comments.

c-blake:
I had some changes in related lines of code that I wanted to make.  So, I just merged this and added my recommended updates to the PR.

